PR #1, opened 2015-05-18
ruby quiz

jamesstaub:


MaxBlaushild:
Score: 4.5/5

Great job! The only reason I'm dinging you is that question 1 is supposed to work for both strings and numbers

jamesstaub:
yes yes, I realized that in review. DIDNT READ INSTRUCTIONS

PR #3, opened 2015-05-18
"Quiz"

LPayyapilli:


MaxBlaushild:
Score: 3/5

The logic in your methods is correct, but you forgot to use your end statements. end statements come at the end of method declarations, at the end of iterations and at the end of if...else statements

PR #4, opened 2015-05-18
Max Ruby Fundamentals Quiz

maxkrieg:


MaxBlaushild:
Score: 4.5/5

Great job. the only reason im dinging you is that the first question should work for both strings and and numbers

PR #5, opened 2015-05-18
quiz

alinaw89:


MaxBlaushild:
Score: 3/5

Great job. The only errors you made were in the first question and with leaving out the `end` statements in your `if...else` statements. The syntax goes as follows:
```ruby
if foo
   bar
else
   baz
end
```

PR #6, opened 2015-05-18
submitting ruby quiz

amygoldberg:


quiz.rb
```diff
@@ -50,6 +62,15 @@ def ac_needed current_temp, ac_working, desired_temp
 
 # your code starts here
 
+def fizzbuzz? num
+  if num % 3 == 0? then
```
MaxBlaushild: You don't need the `?`'s here

MaxBlaushild:
Score: 4/5

Great job. The only errors you made were adding the `?`'s leaving out the end statements in your if...else statements. The syntax goes as follows:
```ruby
if foo
   bar
else
   baz
end
```

PR #7, opened 2015-05-18
quiz.

breduffy:


MaxBlaushild:
Score: 4/5

Great job. The only errors you made were leaving out the `end` statements in your `if...else` statements. The syntax goes as follows:

```javascript
if foo
   bar
else
   baz
end
```

PR #8, opened 2015-05-18
well you know!!!

aaronw3373:


quiz.rb
```diff
@@ -50,6 +61,20 @@ def ac_needed current_temp, ac_working, desired_temp
 
 # your code starts here
 
+def fizz_buzz number
+  num = 0
+  while num <= number
```
MaxBlaushild: no need to iterate here

MaxBlaushild:
Score: 5/5

Great job. Looks like you went above and beyond and printed the fizzbuzz values of the (1..num) range

PR #9, opened 2015-05-18
Quiz Over.

astoellis:


quiz.rb
```diff
@@ -7,7 +7,13 @@
 # it should return false if age is less than minimum_age
 
 def is_of_age? age, minimum_age
-  # your code goes here
+  puts age
+  puts minimum_age
+    if age.to_i <= minimum_age.to_i then
```
MaxBlaushild: you dont need the `if..else` statement here. `age.to_i <= minimum_age.to_i` will evaluate to true or false. also, as you probably noticed, this is backwards

quiz.rb
```diff
@@ -49,6 +66,25 @@ def ac_needed current_temp, ac_working, desired_temp
 # note the space in "fizz buzz"
 
 # your code starts here
+def fizz_buzz num
+  counter = 0
+  while counter <= num
```
MaxBlaushild: no need to enumerate here

MaxBlaushild:
Score: 3.5/5

Your syntax is on point, you just made the aforementioned errors

PR #10, opened 2015-05-18
F#%king fizzbuzz

Peter-White:


quiz.rb
```diff
@@ -48,7 +69,20 @@ def ac_needed current_temp, ac_working, desired_temp
 # and "fizz buzz" if the number is evenly divisible by both 3 and 5
 # note the space in "fizz buzz"
 
-# your code starts here
+def fizzbuzz num
+
+  if num % 5 === 0 && num % 3 === 0
```
MaxBlaushild: The ruby `===` operator is different than the javascript `===` operator. when working with ruby, use `==` instead

MaxBlaushild:
Score: 4/5

Good job. Just remember the aforementioned note. Also, in ruby, you dont need to explicitly type `return` to return a value

PR #11, opened 2015-05-18
quizzle

toomanyadverbs:


quiz.rb
```diff
@@ -48,7 +55,17 @@ def ac_needed current_temp, ac_working, desired_temp
 # and "fizz buzz" if the number is evenly divisible by both 3 and 5
 # note the space in "fizz buzz"
 
-# your code starts here
+def fizz_buzz num
+  if num % 3 == 0 then
+    $stdout.puts "fizz"
+  elsif num % 5 == 0 then
+    $stdout.puts "buzz"
+  elsif num % 15 == 0 then
```
MaxBlaushild: being divisible by 15 is not the same as being divisible by both 3 and 5

MaxBlaushild:
Score: 4.5/5

Great job.

In this instance, you don't need $stdout do print to the console. You can just use `puts`

PR #12, opened 2015-05-18
quizzer

Fitfouraking:


MaxBlaushild:
Score: 4/5

Good job. Just two notes of syntax:

1. you need to close your conditional statements with `end` aka:
```javascript
if foo
   do a thing
else
   do another thing
end
```
2. on question 2, you need to pass the parameters of `okay_to_drink` into `is_of_age?`

PR #13, opened 2015-05-18
my quizzzz

tuskerette:


quiz.rb
```diff
@@ -19,7 +20,10 @@ def is_of_age? age, minimum_age
 # Remember p, print, puts don't return their arguments
 
 def okay_to_drink age, drinking_age
-  # your code goes here
+  if !is_of_age?
```
MaxBlaushild: you need to pass the arguments of `okay_to_drink` into `is_of_age?` here

MaxBlaushild:
Score: 4/5

Good job. You had the syntax almost correct, but you need to close out each of your conditionals with an `end` statement:

```javascript
if foo
  something happens
else
  something else happens
end
```

PR #14, opened 2015-05-18
quazz

Hogue:


quiz.rb
```diff
@@ -19,7 +23,10 @@ def is_of_age? age, minimum_age
 # Remember p, print, puts don't return their arguments
 
 def okay_to_drink age, drinking_age
-  # your code goes here
+  unless is_of_age?  then
```
MaxBlaushild: you need to pass in the arguements of `okay_to_drink` into `is_of_age?`

MaxBlaushild:
Score: 4/5

Good start into rubyland. just a few quick hits:
- its `elsif` not `elseif`
- `then` statements dont hurt but arent needed

other than that, good job

PR #15, opened 2015-05-18
changes to quiz

csadusky:


quiz.rb
```diff
@@ -19,10 +19,14 @@ def is_of_age? age, minimum_age
 # Remember p, print, puts don't return their arguments
 
 def okay_to_drink age, drinking_age
-  # your code goes here
+  is_of_age? age drinking_age
```
MaxBlaushild: Your thinking is right here, but you need to wrap the logic in a conditional statement i.e. 

```javascript
if is_of_age? age, drinking_age
   do something
else
   do something
end
```

quiz.rb
```diff
@@ -36,7 +40,12 @@ def okay_to_drink age, drinking_age
 # below the desired temperature the method should return
 # "Fix the A/C whenever you have the chance. It's cool."
 def ac_needed current_temp, ac_working, desired_temp
-  # your code goes here
+  ac_working == true, current_temp > desired_temp
+  return "Turn on the AC please"
+  ac_working == false, current_temp > desired_temp
```
MaxBlaushild: Thinking here is correct too, but you need to use `&&` instead of commas. `&&` will make the block of code underneath fire only if both conditions are true

MaxBlaushild:
Score: 2.5/5

Your thinking about these problems the right way, so don't sweat the syntax errors yet. The syntax will come with practice

PR #16, opened 2015-05-18
quiz

alik864:


quiz.rb
```diff
@@ -6,8 +6,11 @@
 # it should accept strings or numbers in its arguments
 # it should return false if age is less than minimum_age
 
+
+
 def is_of_age? age, minimum_age
-  # your code goes here
+ if age < minimum age
```
MaxBlaushild: Since you're using this method in later methods, you want it to return a value instead of printing a value to the console. The logic is right tho

MaxBlaushild:
Score: 1.5/5

Switching to a new language is tricky, but you'll get the hang of it soon. Just remember to be conscious of what your methods return

PR #18, opened 2015-05-18
quiz

christobeck:


quiz.rb
```diff
@@ -7,7 +7,8 @@
 # it should return false if age is less than minimum_age
 
 def is_of_age? age, minimum_age
-  # your code goes here
+  minimum_age >= 21
```
MaxBlaushild: You have the right idea, but the age of the person and the minimum age are both parameters or the method, so you would do ` age.to_i >= minimum_age.to_i`

MaxBlaushild:
Score: 1.5/5

Your syntax is on point. Just make sure you close your `if....else` statements with an end

PR #21, opened 2015-07-14
Had some trouble on the last one, but the rest felt good.

joebegley15:


challenge.rb
```diff
@@ -6,7 +6,7 @@
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+  age > minimum_age
```
MattB-GA-Boston: Should probably have been `>=`, but otherwise spot on.

challenge.rb
```diff
@@ -22,7 +22,7 @@ def is_of_age? age, minimum_age
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  age > drinking_age
```
MattB-GA-Boston: You were supposed to solve this by invoking `is_of_age?`; also, your code doesn't return the string that was asked for.

challenge.rb
```diff
@@ -45,7 +45,15 @@ def okay_to_drink? age, drinking_age
...
-  # Replace this comment with your code.
+  if ((current_temp > desired_temp) && (ac_working == true))
+    p "Turn on the AC."
+  elsif ((current_temp <= desired_temp) && (ac_working == true))
+    p "Just right!"
+  elsif ((current_temp > desired_temp) && (ac_working != true))
+    p "Fix the AC now! It's hot!"
+  else
+    p "Fix the AC whenever you have the chance. It's cool."
+
```
MattB-GA-Boston: Should have been `return` rather than `p` - maybe we want to use the result of `ac_monitor` to dive another program!

challenge.rb
```diff
@@ -57,7 +65,22 @@ def ac_monitor current_temp, ac_working, desired_temp
 # - return "fizzbuzz" if `num` is evenly divisible by both 3 and 5
 # - return the value of `num` if `num` isn't divisible by either 3 or 5
 
-# Replace this comment with your code.
+min = #choose
```
MattB-GA-Boston: For this question, you just needed to write a method to calculate the response, e.g. "fizz", "buzz", "fizzbuzz", "other". No need to loop.

challenge.rb
```diff
@@ -67,6 +90,20 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+        puts "fizzbuzz"
+      elsif counter%3 == 0
+        puts "fizz"
+      elsif counter%5 == 0
+        puts "fizzbuzz"
+      else
+          puts counter
+      end
+  end
+end
```
MattB-GA-Boston: Unfortunately, this solution won't work unless you start off by setting counter to min. Also, it doesn't call another method, `fizz_buzz_calculator`, to figure out the right response.

MattB-GA-Boston:
2 / 5 (-1 on Q2, -0.5 on Q3, -1 on Q4, -0.5 on Q5)

Joe, would you be willing to try taking this quiz again on Thursday and resubmitting it? I'd like to see how you fare with it after a few more days of Ruby.

PR #26, opened 2015-07-14
completes ruby assessment

lparvinsmith:


challenge.rb
```diff
@@ -6,9 +6,13 @@
...
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+  if age.to_i >= minimum_age.to_i
+    return true
+  else
+    return false
+  end
 end
-
+#tested
```
MattB-GA-Boston: This is correct! However, I want to point out that saying
```
if some_condition
  return true
else 
  return false
end
```
is kind of redundant. Why? If the condition is true, we should return `true`, and if the condition is false, we should return `false`. So why not just return the condition?
```
  age.to_i >= minimum_age.to_i
end
```

challenge.rb
```diff
@@ -22,9 +26,13 @@ def is_of_age? age, minimum_age
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  if is_of_age?(age, drinking_age)
+    puts "come on in"
+  else
+    puts "I'm sorry, #{age} isn't old enough to drink. The minimum age is #{drinking_age}."
+  end
```
MattB-GA-Boston: Close - this method should return those strings, not print them.

challenge.rb
```diff
@@ -45,9 +53,17 @@ def okay_to_drink? age, drinking_age
...
-  # Replace this comment with your code.
+  if ac_working && current_temp > desired_temp
+    puts "turn on the AC"
+  elsif ac_working && current_temp <= desired_temp
+    puts "just right"
+  elsif current_temp > desired_temp
+    puts "fix AC now! too hot!!"
+  else
+    puts "fix the AC whenever you have the chance. It's cool"
+  end
```
MattB-GA-Boston: Good use of logic. Again, though, this method should be returning a value rather than printing one.

challenge.rb
```diff
@@ -57,8 +73,18 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+  if num %15 ==0
+    puts "fizzbuzz"
+  elsif num % 3 == 0
+    puts "fizz"
+  elsif num % 5 == 0
+    puts "buzz"
+  else
+    puts num
+  end
+end
```
MattB-GA-Boston: See above re: printing and returning.

challenge.rb
```diff
@@ -67,6 +93,13 @@ def ac_monitor current_temp, ac_working, desired_temp
 # and print out the result of `fizz_buzz_calculator` for that number.
 # This question will be graded independently of question four.
 
-# Replace this comment with your code.
+def fizz_buzz min, max
+  i = min
+  while i <= max do
+    fizz_buzz_calculator(i)
```
MattB-GA-Boston: And this is why it matters. If you don't return anything from your method, you can't plug that method's output into any other methods.

MattB-GA-Boston:
4 / 5 (-0.25 on Q2, 3, 4, and 5)

Watch out for returning values vs printing them. Otherwise, nice job.

PR #27, opened 2015-07-14
Completed challenge ruby basics

lekova:


challenge.rb
```diff
@@ -6,7 +6,7 @@
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+  return age < minimum_age
```
MattB-GA-Boston: Should have been _greater than_ minimum age; however, I won't take points off because the question is a bit ambiguous.

challenge.rb
```diff
@@ -22,7 +22,11 @@ def is_of_age? age, minimum_age
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  if is_of_age(age, drinking_age)
+     "I'm sorry, #{age} isn't old enough to drink. The minimum age is #{drinking_age}."
+  else
+    puts "Come on in."
+  end
```
MattB-GA-Boston: Based on your answer to Q1, this is correct - nearly. However, the question also says that the string should be _returned_, not printed.

challenge.rb
```diff
@@ -45,10 +49,17 @@ def okay_to_drink? age, drinking_age
...
-  # Replace this comment with your code.
+  if ac_working && current_temp.to_i > desired_temp.to_i
+    return "Turn on the AC."
+  elsif ac_working && current_temp == desired_temp
+    return "Just right!"
+  elsif !ac_working &&  current_temp > desired_temp
+    return "Fix the AC now! It's hot!"
+  else
+    "Fix the AC whenever you have the chance. It's cool."
+  end
```
MattB-GA-Boston: Correct! In fact, you don't need the other three `return`s either; in any of those cases, the string would be the last line of code evaluated.

challenge.rb
```diff
@@ -57,8 +68,22 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+    elsif num % 3 == 0
+      return 'fizz'
+    elsif num % 5 == 0
+      return 'buzz'
+    else
+      return num
+    end
+    i +=1
+  end
+end
```
MattB-GA-Boston: Not quite - this method was only supposed to figure out the result, not loop through all the numbers (also, the `return` will cause your loop to stop cold after the first run through). 

Since your logic was correct, though, I'll give you partial credit on this one.

challenge.rb
```diff
@@ -67,6 +92,19 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+  #   p fizz_buzz_calculator current_num
+  # end
+
+  # option 2
+  i = min
+  while i >= min && i <= max
+    p fizz_buzz_calculator i
+    i += 1
+  end
+end
```
MattB-GA-Boston: Correct! You could also have written this using `.each` : 
```
(min..max).each { |num| puts fizz_buzz_calculator(num) }
```

MattB-GA-Boston:
4.25 / 5

Great work, Lina!

PR #30, opened 2015-07-14
Submits assessment

aemassa:
Aw man! I'll finish this later :)

### DIFF
--- a/challenge.rb
+++ b/challenge.rb
@@ -6,7 +6,13 @@
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+  if age < minimum_age
+    return false
+    puts "false"
+  else
+    return true
+    puts "true"
+  end
 end
 
 
--- a/challenge.rb
+++ b/challenge.rb
@@ -6,12 +6,10 @@
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  if age < minimum_age
+  if age.to_i < minimum_age.to_i
     return false
-    puts "false"
   else
     return true
-    puts "true"
   end
 end
 
@@ -28,7 +26,11 @@ end
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  if (is_of_age? age, drinking_age) == true
+      return "Come on in."
+  else
+      return "I'm sorry, #{age} isn't old enough to enter. The minimum age is #{drinking_age}."
+  end
 end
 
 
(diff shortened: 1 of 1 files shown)


challenge.rb
```diff
@@ -22,7 +26,11 @@ def is_of_age? age, minimum_age
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  if (is_of_age? age, drinking_age) == true
+      return "Come on in."
+  else
+      return "I'm sorry, #{age} isn't old enough to enter. The minimum age is #{drinking_age}."
+  end
```
MattB-GA-Boston: All correct! However, it could be a little tighter. For instance, the pattern `x == true` is redundant, since the expression is only ever true when `x` is true, and only ever false when `x` is false. In your code above, you could have just written
```
    if (is_of_age? age, drinking_age)
```

challenge.rb
```diff
@@ -45,7 +53,18 @@ def okay_to_drink? age, drinking_age
...
+  if current_temp <= desired_temp && ac_working == true
+    return "Just right!"
+  end
+  if current_temp > desired_temp && ac_working == false
+    return "Fix the AC now! It's hot!"
+  end
+  if current_temp <= desired_temp && ac_working == false
+    return "Fix the AC whenever you have the chance. It's cool."
+  end
```
MattB-GA-Boston: See previous comment about `== true`. `x == false` is similarly redundant, since it will always be equal to `!x`. 

This set of conditions could also have been written a bit more tersely. Ultimately, we only need to perform two logical tests to break apart all four possible cases: (1) is the AC working, and (2) is it too hot. Imagine if we wrote the following:
```
if ac_working
  if current_temp > desired_temp
    "Turn on the AC"
  else
    "Just right."
  end
else 
  if current_temp > desired_temp
    "Fix the AC now! It's hot!"
  else 
    "Fix the AC whenever you have the chance. It's cool."
  end
end
```

challenge.rb
```diff
@@ -57,7 +76,17 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+  if num % 3 == 0 && num % 5 == 0
+    return "fizzbuzz"
+  elsif num % 3 == 0
+    return "fizz"
+  elsif num % 5 == 0
+    return "buzz"
+  else
+    return num
+  end
+end
```
MattB-GA-Boston: Great! One small nit - the `return` statements aren't needed here, because a Ruby method or block will return the value of the last evaluated expression by default.

challenge.rb
```diff
@@ -67,6 +96,11 @@ def ac_monitor current_temp, ac_working, desired_temp
 # and print out the result of `fizz_buzz_calculator` for that number.
 # This question will be graded independently of question four.
 
-# Replace this comment with your code.
-
-
+def fizz_buzz min, max
+  i = min
+  while i <= max do
+    fizz_buzz_calculator(i)
```
MattB-GA-Boston: This line's not necessary.

challenge.rb
```diff
@@ -67,6 +96,11 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+    puts fizz_buzz_calculator(i)
+    i += 1
+  end
+end
```
MattB-GA-Boston: Spot on! 

Here's a more terse version, leveraging the power or Ruby ranges. 
```
def fizz_buzz min, max
  (min..max).each do |num|
    puts fizz_buzz_calculator(num)
  end
end
```

PR #32, opened 2015-07-14
Finished test thingy

payne-chris-r:


challenge.rb
```diff
@@ -5,10 +5,15 @@
...
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
-def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+def is_of_age?(age, minimum_age)
+  if age < minimum_age
+    false
+  else
+    true
+  end
```
MattB-GA-Boston: Correct. A more terse way to write this would be to say
```
def is_of_age?
    age >= minimum_age
end
```

challenge.rb
```diff
@@ -5,10 +5,15 @@
...
 end
 
+puts is_of_age?(20, 21)
```
MattB-GA-Boston: Testing your work? I approve!

challenge.rb
```diff
@@ -21,11 +26,15 @@ def is_of_age? age, minimum_age
...
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
-def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+def okay_to_drink?(age, drinking_age)
+  if (is_of_age?(age, drinking_age) == true)
+    puts "Come on in."
+  else
+    puts "I'm sorry, #{age} isn't old enough to dwwink beews. The min age is #{drinking_age}."
+  end
```
MattB-GA-Boston: Supposed to return instead of print, but otherwise correct. But per my comment in Q1, this could also have been written a bit more tersely. The statement `x == true` is redundant. If `x` is true, the expression evaluates to `true`, and if `x` is false, the expression evaluates to `false`. In short, there's no reason why we couldn't write `x` instead of `x == true`. Or, in your case, just writing `if is_of_age?(age, drinking_age)`

challenge.rb
```diff
@@ -44,10 +53,23 @@ def okay_to_drink? age, drinking_age
...
+    else
+      "Fix the AC whenever."
+    end
+  else
+    if(ac_working)
+      "Turn on the AC."
+    else
+      "Fix the AC now! It's hot!"
+    end
+  end
```
MattB-GA-Boston: Nice. Logical expressions would also have been appropriate here.

challenge.rb
```diff
@@ -57,8 +79,21 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+      "fizzbuzz"
+    else
+      "buzz"
+    end
+  elsif (num % 3 == 0)
+    "fizz"
+  else
+    "#{num}"
+  end
+end
```
MattB-GA-Boston: Great. Spot on.

challenge.rb
```diff
@@ -67,6 +102,8 @@ def ac_monitor current_temp, ac_working, desired_temp
 # and print out the result of `fizz_buzz_calculator` for that number.
 # This question will be graded independently of question four.
 
-# Replace this comment with your code.
-
+def fizz_buzz(min,max)
+  puts (min..max).map { |i| fizz_buzz_calculator(i) }
+end
```
MattB-GA-Boston: Not quite the spec - we were supposed to print out each number separately. I probably would have used `each` to accomplish that, with a `puts` inside the block. Still, you're pretty close to the mark.

MattB-GA-Boston:
4 / 5 (-0.5 on Q2, -0.5 on Q5)

Very nice job. I recommend doing some exercises on CodeWars to make your Ruby even tighter.

PR #33, opened 2015-07-14
Completes challenge.rb.

kristenlk:


challenge.rb
```diff
@@ -1,12 +1,17 @@
...
 
+require 'pry'
 
 # Question 1
 # Complete the method `is_of_age?`. It should accept either strings or numbers in its arguments,
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+  if (age.to_s >= minimum_age.to_s) || (age.to_i >= minimum_age.to_i)
```
MattB-GA-Boston: Not necessary to compare the string case.

challenge.rb
```diff
@@ -1,12 +1,17 @@
...
+    return true
+  else
+    return false
+  end
```
MattB-GA-Boston: This is correct, but could have been written more tersely.
```
   age >= minimum_age
end
```
Do you see how these work out to be the same?

challenge.rb
```diff
@@ -21,8 +26,13 @@ def is_of_age? age, minimum_age
 # This question will be graded independently of question one.
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
+
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  if is_of_age?(age, drinking_age) == true
```
MattB-GA-Boston: The idiom `x == true` is redundant; if the expression evaluates as true, it must be because `x` is true (and vice versa). In other words, it's the same as just writing `x` - or, in this case, 
`if is_of_age?(age, drinking_age)`

challenge.rb
```diff
@@ -21,8 +26,13 @@ def is_of_age? age, minimum_age
...
+    return 'Come on in.'
+  else
+    return "I'm sorry, #{age} isn't old enough to drink. The minimum age is #{drinking_age}."
```
MattB-GA-Boston: Nice job with the string interpolation.

challenge.rb
```diff
@@ -45,10 +55,19 @@ def okay_to_drink? age, drinking_age
...
-  # Replace this comment with your code.
+  if ac_working == true && (current_temp > desired_temp)
+    return "Turn on the AC."
+  elsif ac_working == true && (current_temp <= desired_temp)
+    return "Just right!"
+  elsif ac_working == false && (current_temp > desired_temp)
+    return "Fix the AC now! It's hot!"
+  elsif ac_working == false && (current_temp <= desired_temp)
+    return "Fix the AC whenever you have the chance. It's cool."
+  end
```
MattB-GA-Boston: Correct, but see my above comment about redundancy. You can just write `ac_working` or `!ac_working` to indicate whether the value of `ac_working` is `true` or `false`.

challenge.rb
```diff
@@ -57,8 +76,17 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+  if num % 15 == 0
+    return 'fizzbuzz'
+  elsif num % 3 == 0
+    return 'fizz'
+  elsif num % 5 == 0
+    return 'buzz'
+  else
+    return num
+  end
+end
```
MattB-GA-Boston: Bingo! Nice one.

challenge.rb
```diff
@@ -67,6 +95,14 @@ def ac_monitor current_temp, ac_working, desired_temp
 # and print out the result of `fizz_buzz_calculator` for that number.
 # This question will be graded independently of question four.
 
-# Replace this comment with your code.
 
+# not done-
+def fizz_buzz min max
+  while min < max
+    return fizz_buzz_calculator(min)
```
MattB-GA-Boston: Should be `print` or `puts` in this one - using `return` will break us out of the loop, so we won't finish retrieving the numbers. It also doesn't strictly print...

MattB-GA-Boston:
4.5 / 5

Most of my feedback was stylistic - how to write better code - but the first priority is always having your code be functional, and you succeeded in that. Great job, Kristen!

PR #34, opened 2015-07-14
Time and Test

GHolmes04:
Not enough time to complete Q5 or test other answers

### DIFF
--- a/challenge.rb
+++ b/challenge.rb
@@ -6,7 +6,8 @@
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+ check = (minimun_age.to_i) - age.to_i)
+ check >= 0
 end
 
 
@@ -22,7 +23,7 @@ end
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  check >=0 ? "Come on in" : "I'm sorry " + age + " ins't old enough to drink. The minimun age is " + drinking_age +"."
 end
 
 
@@ -45,7 +46,16 @@ end
 #   "Fix the AC whenever you have the chance. It's cool."
 
 def ac_monitor current_temp, ac_working, desired_temp
-  # Replace this comment with your code.
+  if ((ac_working == true) && (current_temp > desired_temp))
+    puts "Turn on the AC."
+  elsif (ac_working == true) && (current_temp <= desire_temp))
+    puts "Just Right!"
+  elsif ((ac_working == false) && (current_temp <= desire_temp))
+    puts "Fix the AC whenever you have the chance. It's cool"
+  else "Fix the AC now! It's hot!"
+
+
+  end
 end
 
 
@@ -57,7 +67,26 @@ end
 # - return "fizzbuzz" if `num` is evenly divisible by both 3 and 5
 # - return the value of `num` if `num` isn't divisible by either 3 or 5
 
-# Replace this comment with your code.
+ def fizz_buzz_calculator(num
+    if num%5 == 0 && num%3 == 0
+      puts "Fizzbuzz"
+
+    elsif num%3 == 0
+      puts 'Fizz'
+
+    elsif num%5 == 0
+      puts 'Buzz'
+
+    else
+      puts i
+
+    end
+    i += 1
+  end
+
+  return list
+end
+
 
 
 # Question 5 : "Return of the FizzBuzz"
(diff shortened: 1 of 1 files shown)


challenge.rb
```diff
@@ -6,7 +6,8 @@
 # and should return either `true` or `false` based on whether `age` is less than `minimum_age`.
 
 def is_of_age? age, minimum_age
-  # Replace this comment with your code.
+ check = (minimun_age.to_i) - age.to_i)
+ check >= 0
```
MattB-GA-Boston: Should be `true` if age is _greater_ than _minimum age_, but the question is worded a little ambiguously, so I won't take points off.

challenge.rb
```diff
@@ -22,7 +23,7 @@ def is_of_age? age, minimum_age
 # Remember: `p`, `print`, and `puts` don't return their arguments.
 
 def okay_to_drink? age, drinking_age
-  # Replace this comment with your code.
+  check >=0 ? "Come on in" : "I'm sorry " + age + " ins't old enough to drink. The minimun age is " + drinking_age +"."
```
MattB-GA-Boston: The ternary is appropriate, but the string interpolation is off. Unlike JavaScript, Ruby does not implicitly convert a number to a string when the two are added together. We need to either explicitly convert the numbers to strings using `.to_s` or use the built-in string interpolation syntax `"blah blah #{interpolated_value}"`

challenge.rb
```diff
@@ -45,7 +46,16 @@ def okay_to_drink? age, drinking_age
...
 
 def ac_monitor current_temp, ac_working, desired_temp
-  # Replace this comment with your code.
+  if ((ac_working == true) && (current_temp > desired_temp))
+    puts "Turn on the AC."
+  elsif (ac_working == true) && (current_temp <= desire_temp))
+    puts "Just Right!"
+  elsif ((ac_working == false) && (current_temp <= desire_temp))
+    puts "Fix the AC whenever you have the chance. It's cool"
+  else "Fix the AC now! It's hot!"
```
MattB-GA-Boston: Should be returning values, not printing them. Otherwise right (apart from some typos).

challenge.rb
```diff
@@ -57,7 +67,26 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+      puts 'Buzz'
+
+    else
+      puts i
+
+    end
+    i += 1
+  end
+
+  return list
```
MattB-GA-Boston: This method should be returning one of the four values; there's no need to iterate or generate a list. Your fizzbuzz logic is right, though.

challenge.rb
```diff
@@ -67,6 +96,33 @@ def ac_monitor current_temp, ac_working, desired_temp
 # and print out the result of `fizz_buzz_calculator` for that number.
 # This question will be graded independently of question four.
 
-# Replace this comment with your code.
+def fizzbuzz_calc (min max)
+  list = {
+  'Fizz' => [],
+  'Buzz' => [],
+  'FizzBuzz' =>[],
+  'other' => []}
```
MattB-GA-Boston: This was from another exercise - not what we're asking here.

challenge.rb
```diff
@@ -67,6 +96,33 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+
+  i = 1
+  while i <= max && i > max
```
MattB-GA-Boston: This statement will always be true; consequently, we will never leave the `while` loop.

challenge.rb
```diff
@@ -67,6 +96,33 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+      puts 'Fizz'
+      list['Fizz'].push(i)
+    elsif i%5 == 0
+      puts 'Buzz'
+      list['Buzz'].push(i)
+    else
+      puts i
+      list['other'].push(i)
+    end
+    i += 1
```
MattB-GA-Boston: Technically, this solution might work (though it contains the unnecessary `list` object). However, we did also say that you should use the `fizz_buzz_calculator` method from Q4 as part of your solution.

challenge.rb
```diff
@@ -67,6 +96,33 @@ def ac_monitor current_temp, ac_working, desired_temp
...
+  end
+
+  return list
+end
```
MattB-GA-Boston: Assuming that `fizz_buzz_calculator` works, here's as possible solution for Q5.

```
def fizz_buzz (min, max)
  (min..max).each {|num| 
     puts fizz_buzz_calculator(num)
  }
end
```

MattB-GA-Boston:
3.25 / 5 (-0.5 on Q2, -0.25 on Q3, -0.5 on Q4, -0.5 on Q5)

Not bad, Greg! I think you've got this - just try to relax a bit and slow down before jumping right into the code. Map it out in English first, and then translate it into Ruby/JS.

PR #35, opened 2015-07-16
Corrected Errors 

GHolmes04:
Hi Matt thanks for the feed back 

I used my 1 on 1 time to go over the past two assessments here is the quiz completed correctly. Just figured I would send it over. No need to look at it if you don't want to.


MattB-GA-Boston:
Nice! Looks good, Chris!

PR #43, opened 2016-02-11
Ruby diagnostic response

seatuna:


lib/diagnostic.rb
```diff
@@ -40,7 +40,9 @@
 def method_name
 end
 
-Diagnostic.defined_method = # call your method here
+Diagnostic.defined_method = def sum_times_two (a, b)
+                              2 * (a + b)
+                            end
```
J-Weeks: define then call
```

def add_and_multiply(a, b)
  2 * (a + b)
end

Diagnostic.defined_method = add_and_multiply(3, 4)
```

lib/diagnostic.rb
```diff
@@ -49,31 +51,33 @@ def method_name
 
 # define your method here
 
-Diagnostic.predicate = # call your method here
+Diagnostic.predicate = def odd? (number)
+                          number.odd?
+                      end
 
 # Numbers in Ruby
```
J-Weeks: define then call

J-Weeks:
14/16.  Good job!  Keep up the good work.

lib/diagnostic.rb
seatuna: I see. I just realized I misread the question / maybe wasn't reading!

PR #44, opened 2016-02-11
diagnostic response

MFBrewster:


lib/diagnostic.rb
```diff
@@ -37,43 +37,46 @@
 # of multiplying by two the result of adding `a` and `b`. Name the method
 # conventionally.
 
-def method_name
+def twice_product(a, b)
+  (a + b) * 2
 end
 
-Diagnostic.defined_method = # call your method here
+Diagnostic.defined_method = twice_product(num1, num2)
```
J-Weeks: call it with numbers, but I get what you want to do.

lib/diagnostic.rb
```diff
@@ -37,43 +37,46 @@
...
 # Define a method that checks whether a number passed in is odd. Name it
 # conventionally.
 
-# define your method here
+def odd?(num)
+  num % 2 == 1
+end
 
-Diagnostic.predicate = # call your method here
+Diagnostic.predicate = 3.odd?
```
J-Weeks: so .odd? is already a method. be careful with this.  I see what you're trying to do though.

lib/diagnostic.rb
```diff
@@ -37,43 +37,46 @@
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = '1.001.class #returns float'
```
J-Weeks: Float

J-Weeks:
15.5/16.  Good job!

PR #47, opened 2016-02-11
ruby diagnostic

msyao-github:


lib/diagnostic.rb
```diff
@@ -5,28 +5,34 @@
...
+Diagnostic.script = 'pry'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
+pry
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'name = "star wars episode viii"'
```
J-Weeks: need snake case: 
Diagnostic.variable = 'star_wars_episode_viii'

lib/diagnostic.rb
```diff
@@ -5,28 +5,34 @@
...
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+name = "star wars episode viii"
+puts name
+
+Diagnostic.variable =  "character = #{'Jar Jar Binks'}"
```
J-Weeks: you only want to interpolate one thing and also you seem to be assigning it to a variable. this will blow up.

right answer : Diagnostic.interpolation = "#{character} is really a Sith Lord."

lib/diagnostic.rb
```diff
@@ -37,9 +43,13 @@
...
 
-def method_name
+
+def mult_add(a, b)
+  (a+b)*2
 end
 
+puts mult_add(a,b)
+
 Diagnostic.defined_method = # call your method here
```
J-Weeks: looking for something like this: Diagnostic.defined_method = add_and_multiply(3, 4)

lib/diagnostic.rb
```diff
@@ -49,6 +59,15 @@ def method_name
...
+def odd(num)
+   if num % 2 == 1
+      return "#{numb} is odd"
+   end
+   if num % 2 == 0
+      return "#{num} is even"
+  end
+end
+
 Diagnostic.predicate = # call your method here
```
J-Weeks: numb vs num, check your variable names.  Also you didn't call it.  Why two separate if statements?
something like this:
```
def odd?(number)
  number % 2 == 1
end

```

J-Weeks:
9/16.  I think you need some ruby review.  Try some code wars and ask for help if you feel you need it.  I'm not sure if this was due to time or needing review.

PR #48, opened 2016-02-11
wrote answers

Knguyen21:


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
J-Weeks: this will not work, need snake case
 Diagnostic.variable = 'star_wars_episode_viii'

lib/diagnostic.rb
```diff
@@ -40,7 +40,8 @@
 def method_name
 end
 
-Diagnostic.defined_method = # call your method here
+Diagnostic.defined_method = (a, b).reduce { |a, b| (a + b) * 2}
+
```
J-Weeks: Dont over think it
```
def add_and_multiply(a, b)
  2 * (a + b)
end
```

lib/diagnostic.rb
```diff
@@ -49,31 +50,31 @@ def method_name
 
 # define your method here
 
-Diagnostic.predicate = # call your method here
+Diagnostic.predicate = odd?
```
J-Weeks:
```
def odd?(number)
  number % 2 == 1
end

```

lib/diagnostic.rb
```diff
@@ -49,31 +50,31 @@ def method_name
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
```
J-Weeks: caps, Float

J-Weeks:
12.5/16.  Good job.  Don't over think it.  I feel that you grasp these concepts but sometimes get into your own head.

lib/diagnostic.rb
Knguyen21: opps.. I thought you wanted to set the variable value to 'star wars episode viii' . I didn't read the question correctly

lib/diagnostic.rb
Knguyen21: Okay.. I didn't read the question fully. I didn't know we were suppose to write a method out.

lib/diagnostic.rb
Knguyen21: ohhh.. OKay.  Yeah, I'm definitely overthinking

PR #58, opened 2016-04-13
Diagnostic responses

cuprous:


lib/diagnostic.rb
```diff
@@ -115,10 +117,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = "person = {first_name: \"Zachary\", age: 29}"
```
J-Weeks: person = {first_name: "Jason", age: 27}'
Why the added breaks?

lib/diagnostic.rb
```diff
@@ -115,10 +117,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = "person.default = \"\#{person[:first_name]}"
```
J-Weeks: same here

lib/diagnostic.rb
cuprous: Are you referring to the escaped quotes? If so, rspec broke without them.

J-Weeks:
17/17. Good job.

PR #63, opened 2016-04-13
Response to ruby diagnostic.

natdjerf:


lib/diagnostic.rb
```diff
@@ -29,40 +29,41 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
 
 # Define a predicate method in Ruby
 #
 # Define a method that checks whether a number passed in is odd. Name it
 # conventionally.
 
 # define your method here
+odd_sum.range.reduce { |a, e| e.odd? ? a + e : a }
```
J-Weeks: you don't need to reduce here.  This is an easier way:
```rb
def odd?(number)
  number % 2 == 1
end
```

J-Weeks:
17/17. Great job, keep up the good work.

PR #72, opened 2016-04-13
Response

Jcornmanhomonoff:


lib/diagnostic.rb
```diff
@@ -38,24 +38,27 @@
 
 # define your method here
 
-Diagnostic.predicate = # call your method here
```
laurenfazah: Correct in using odd? for this, but we were asking you to define your own method that would check for this, such as:

```
def odd?(number)
  number % 2 == 1
end

```

lib/diagnostic.rb
```diff
@@ -78,18 +81,18 @@
 
 # Predict what value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
```
laurenfazah: Close, just checking if `batman == true`, so would return `"The Dark Night"`

lib/diagnostic.rb
```diff
@@ -100,13 +103,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
```
laurenfazah: You'd want to use `.pop` here to remove from the end of an array. `arr.pop 2` would then remove the last 2 elements from `arr`

lib/diagnostic.rb
```diff
@@ -100,13 +103,13 @@
...
+Diagnostic.remove_from_array = arr.delete_at(-1, -2)
 
 # What would `arr` equal after:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
```
laurenfazah: Adding 99 to the index at `arr.length +  2` would give you `[12, 34, 56, 67, nil, nil, 99]`

lib/diagnostic.rb
```diff
@@ -115,7 +118,7 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = person_hash = {age: "", first_name: ""}
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
```
laurenfazah: Answer to the last question: `Diagnostic.hash_default = 'person.default = "#{person[:first_name]} is great"`

laurenfazah:
12/17 - good work!

PR #89, opened 2016-08-03
David Powell, diagnostic

dvdpwll:
ansered questions.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -100,13 +103,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = arr[0,2]
```
laurenfazah: This would grab the first two elements. What you'd want to use is `.pop`. So, `arr.pop 2` to `pop` off the last two elements

laurenfazah:
15/16 - good work, David!

PR #90, opened 2016-08-03
Answer questions in Ruby diagnostic

jordanallain:
Fill out answers to questions regarding ruby.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -100,13 +104,14 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = ''
```
laurenfazah: `arr.pop 2` will remove the last two elements from `arr`

lib/diagnostic.rb
```diff
@@ -100,13 +104,14 @@
...
 
 # What would `arr` equal after:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+
+Diagnostic.array_add_to = '[12, 34, 56, 67, , , 99]'
```
laurenfazah: Empty index values would be the case in JS, but Ruby would fill those with `nil`

lib/diagnostic.rb
```diff
@@ -115,10 +120,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = Hash.new('')'
```
laurenfazah: Close, the `person` array should have been instantiated with two keys, `:first_name` and `:age`, so 

`person = {first_name: "Jordan", age: 88}`

laurenfazah:
13/16, good work Jordan!

PR #96, opened 2016-08-03
add answer to ruby diagnostic

EchoXin:


lib/diagnostic.rb
```diff
@@ -38,31 +38,31 @@
 
 # define your method here
 
-Diagnostic.predicate = # call your method here
+Diagnostic.predicate = # &:odd?
```
laurenfazah: This works, but we were looking for a new method defined conventionally for a predicate, such as:

```
def num_odd? (n)
  # something with n
end
```

lib/diagnostic.rb
```diff
@@ -115,7 +115,7 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person={"age":24, "first_name":"ying"}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
```
laurenfazah: hash_default - person.default = `"\#{person[:first_name]}`

laurenfazah:
14/16, good work Ying!

PR #100, opened 2016-09-15
Answer questions

rhjones:
Answer all questions


rebekahheacock:
I'm a little confused about the `array_add_to` question—the spec expects the full original array, plus the new value, but we've used `pop` to remove values from the end of that array. Given that `pop` modifies the original array, why is this not reflected in the answer to `array_add_to`? It seems like it should be:

`[12, 34, nil, nil, 99]`

PR #122, opened 2016-11-09
Complete ruby diagnostic

kiraamaa:
Used Google and class notes to complete

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = "#{character} is really a Sith Lord."
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float' # or number.to_f ?
```
BenGitsCode: just float is all we needed 👌

PR #124, opened 2016-11-09
Respond to questions for ruby-diagnostic

kibs86:
- update diagnostic.rb with responses
- two rake tests failed, but I explained my answers in comments above

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -91,13 +91,16 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+# The rake test failed for below, but this way still works.
+Diagnostic.remove_from_array = 'arr.pop(2)'
```
BenGitsCode: the test looks for `arr.pop 2` but you're dead right, we noticed the test issues yesterday and haven't pushed the changes yet 👌

lib/diagnostic.rb
```diff
@@ -91,13 +91,16 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+# The rake test for this is incorrect.  It uses the original array and not the
+# result from the last question.
+Diagnostic.array_add_to = '[12, 34, nil, nil, 99]'
```
BenGitsCode: Yes. The spec for this question is so wrong it's laughable 😆

PR #125, opened 2016-11-09
Completes diagnostic.rb

paulsevere:
Completes diagnostic.rb


lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr[-2..-1] = [] '
```
BenGitsCode: Unique approach! 👍

PR #126, opened 2016-11-09
Ruby Diagnostic on 11/9/16

Sphilippe7:
Most of the answers (if not all) are incorrect but I did my best

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -1,12 +1,12 @@
 require 'ostruct'
-Diagnostic = OpenStruct.new
 
+Diagnostic = OpenStruct.new
 # Running scripts from the command line
 #
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = './lib/example.rb'
 
 # Using the REPL
 #
@@ -20,7 +20,7 @@ Diagnostic.repl = 'your response here'
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = '@starwarsepisodeviii = star wars episode viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -1,12 +1,12 @@
...
-Diagnostic = OpenStruct.new
 
+Diagnostic = OpenStruct.new
 # Running scripts from the command line
 #
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = './lib/example.rb'
```
BenGitsCode: close! 

`pry OR ruby lib/example.rb`

lib/diagnostic.rb
```diff
@@ -29,14 +29,14 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '@character is really a Sith Lord.'
```
BenGitsCode: Remember how we did this on Monday?

using #{character} you in interpolate the string value into a string.

lib/diagnostic.rb
```diff
@@ -29,14 +29,14 @@
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = ''
```
BenGitsCode: float

lib/diagnostic.rb
```diff
@@ -80,7 +80,7 @@
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = false
```
BenGitsCode: Nice! 👍

lib/diagnostic.rb
```diff
@@ -91,7 +91,7 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.delete[2, 3]'
```
BenGitsCode: Try running that in your repl.

There are a lot of possible answers, `arr.pop 2` would be a simple example.

PR #127, opened 2016-11-09
Answer diagnostic questions

akradin:
Sources used: pry,  rubylearning.com, Ruby documentation for hashes,
  and stack overflow

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative 'lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative 'lib/example.rb'
```
BenGitsCode: `ruby lib/example.rb`

Just like how we used `node <path here>`

akradin:
if we are in pry, can’t we do require_relative as well?

lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
...
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = arr.pop(2)
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '12, 34, nil, nil, 99'
```
BenGitsCode: `'[12, 34, 56, 67, nil, nil, 99]'`

PR #130, opened 2016-11-09
respond to ?s

ioncrash:
response

comfort: 4
clarity: 5


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative 'lib/example.rb'
```
BenGitsCode: `pry` OR `ruby` `lib/example.rb`

PR #131, opened 2016-11-09
Response to ruby diagnostic.

ielole:
Answered questions.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'pry ruby/lib/example.rb'
 
 # Using the REPL
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'pry ruby/lib/example.rb'
```
BenGitsCode: remove the `ruby/` and this is on point.

lib/diagnostic.rb
```diff
@@ -69,18 +69,18 @@
...
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = true
```
BenGitsCode: False.

It never requires an explicit return. That's why the dark knight question doesn't need one and still returns 'the dark knight'

PR #132, opened 2016-11-09
Diagnostic Tues Nov 9

RpRoderick:
Ruby

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby 'lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry' 'lib/example.rb'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
...
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby 'lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry' 'lib/example.rb'
```
BenGitsCode: Just looking for how to open it up, so just `pry` would suffice.

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = "#{character} + 'is really a Sith Lord.''"
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = '1.9 + 3.2.to_f  (a float)'
```
BenGitsCode: float

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = '1.9 + 3.2.to_i '
```
BenGitsCode: fixed number

lib/diagnostic.rb
```diff
@@ -91,7 +91,7 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.delete something, running out of time!'
```
BenGitsCode: you could use delete, among many other things. one of the simplest answers would be 

`arr.pop 2`

PR #133, opened 2016-11-09
ruby diagnostic

akadane87:


lib/diagnostic.rb
```diff
@@ -6,7 +6,8 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+
+Diagnostic.script = pry 'lib/tasks/example.rb'
```
BenGitsCode: where the heck did tasks come from?

lib/diagnostic.rb
```diff
@@ -20,7 +21,7 @@
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "star wars episode viii"
```
BenGitsCode: camel case `star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -29,14 +30,14 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} ANSWER is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'integ'
```
BenGitsCode: float. Integer = 1 float = 1.1

lib/diagnostic.rb
```diff
@@ -91,13 +92,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.delete_list[2,3]'
```
BenGitsCode: `arr.pop 2`

There are other options but your answer won't run in pry.

akadane87:
Ahh tasks was a sub directory of lib but on the same level as the
diagnostic...not its parent

¯\_(ツ)_/¯

PR #138, opened 2016-11-09
ans all questions

jbuzzanco:
answered  all questions to best of my knowledge

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -4,23 +4,25 @@ Diagnostic = OpenStruct.new
 # Running scripts from the command line
 #
 # In a Ruby string, write the command you would use to run a script at
+
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = pry './lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
+
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "star wars epsode vii"
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -4,23 +4,25 @@
 # Running scripts from the command line
 #
 # In a Ruby string, write the command you would use to run a script at
+
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = pry './lib/example.rb'
```
BenGitsCode: Technically this would go back one folder. So if we're in project, and want to access `project/lib/example.rb` we only need `lib/example.rb` this would go up one parent directory first.

lib/diagnostic.rb
```diff
@@ -69,18 +71,19 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'nothing will be returned as there is
+nothing asking to be returned'
```
BenGitsCode: Correct, if this were javascript. Try running this code in pry and see what gets returned ;)

lib/diagnostic.rb
```diff
@@ -91,13 +94,14 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'use delete method to select "56" and "67"'
```
BenGitsCode: You could use the delete method but we were looking for `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -91,13 +94,14 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = 'It would not add 99 because that index does not exist
+at that length'
```
BenGitsCode: we were looking for what that code would result in.

`'[12, 34, 56, 67, nil, nil, 99]'`

Always try running code like this yourself to see the return.

lib/diagnostic.rb
```diff
@@ -106,10 +110,15 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'class Person
+  def initialize(first_name, last_name)
+  @first_name = first_name
+  @last_name = last_name
+  end
+  Person.first_name('leslie').last_name('knope')'
```
BenGitsCode: `person = {first_name: 'Ben', age: 28}`

we were looking for this, does that maks sense?

PR #146, opened 2017-01-26
Update solution for cohort 016

MicFin:
Consider removing excess branches like `015/solution` when you delete `solution` and rename `016/solution` to `solution`.


lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'Fixnum'
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['3.3', '3']
```
BenGitsCode: These are strings, they can't be for the test to pass.

```diff
- Diagnostic.numbers = ['3.3', '3']
+ Diagnostic.numbers = [3.3, 3]
```

lib/diagnostic.rb
```diff
@@ -91,13 +91,15 @@
...
+Diagnostic.remove_from_array = 'arr.pop 2' # could use pop, slice, push etc
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[12, 34, 56, 67, nil, nil, 99]'
+# or (if they're basing it on having popped 2 from the end)
+Diagnostic.array_add_to = '[12, 34, nil, nil, 99]'
```
BenGitsCode: This test doesn't pass, and I can think of two possible ways to fix it:

1. explicitly ask them to use pop, use only that solution and ensure the test checks for that one.
2. Keep this as is, but have the test check for both. I think there's an `or` operator in rspec but I haven't been able to get it to work myself. 

Rspec isn't my forte, so I'm not sure which fix works better, but let's get those tests passing!

BenGitsCode: I think you meant to point this to solution too, but since we don't merge on github, that won't affect anything.

BenGitsCode:
@MicFin I just fixed the first test fixed, but I'm struggling to find the better fix for the add to array test. We'll deliver master today but get someone other than you and I to review and merge this after the above changes are made.

MicFin:
Closing.  `LM01` branch updates are correct https://github.com/ga-wdi-boston/ruby-diagnostic/pull/173.

PR #149, opened 2017-02-02
Add responses to diagnostic

akgoode:
Comfort: 5
Clarity: 5

Used google and Ruby documentation to double check answers


bernardlee:
15/15

PR #151, opened 2017-02-02
Submit Diagnostic

jwaskow:
Comfort: 5
Clarity: 5

I like how this diagnostic was longer and more of a 'quiz' than a 'challenge', it helped me really understand the definitions for ruby terms.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'def "star wars episode viii"'
```
bernardlee: `def` is used for defining functions.  Ruby uses "snake case" as its naming convention, so you would want to name this variable `star_wars_episode_episode_viii`.

lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
...
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[12, 34, nil, 99]'
```
bernardlee: Almost.  There's an extra `nil` in there.  `[12, 34, nil, nil, 99]`.

bernardlee:
13/15

PR #153, opened 2017-02-02
Diagnostic - Ruby

sinyru:
Comfort: 3
Clarity: 4

I used stackoverflow and the require readings on Ruby for this.
I am not sure on how to solve the questions about Hash. If you can let
me know, that would be great. Thanks!

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,22 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+
+Diagnostic.script = "ruby lib/example.rb"
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = "pry"
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "star wars episode viii"
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,22 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = "pry"
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "star wars episode viii"
```
bernardlee: Ruby uses "snake case" for its naming convention so this should be `star_wars_episode_viii`.

lib/diagnostic.rb
```diff
@@ -91,13 +92,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = '2.times {arr.pop}'
```
bernardlee: Another way is `arr.pop(2)`.

lib/diagnostic.rb
```diff
@@ -106,10 +107,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = {age:27,first_name:dave}'
```
bernardlee: This will cause an error since `dave` is not defined.

lib/diagnostic.rb
```diff
@@ -106,10 +107,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person.default = person[:first_name] + message'
```
bernardlee: This will cause an error since `message` is not defined.

bernardlee:
12/15

PR #154, opened 2017-02-02
rub y diagnostic

ggwilliams:
-answered questions

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby lib/example.rb
 
-# Using the REPL
+# Using the REP
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = print "star wars episode viii"
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = print "star wars episode viii"
```
bernardlee: We want to know what you would name a variable using Ruby's naming convention (snake case).  `star_wars_episode_viii`.

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = print '#{character} is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = to_f
```
bernardlee: `to_f` will attempt to convert the object that it's invoked upon into a `Float`.  The correct answer is `Float`.

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'to_i'
```
bernardlee: `to_i` will attempt to convert the object that it's invoked upon into an integer.  e.g., `'8'.to_i` will attempt to convert the string `'8'` into the integer `8`.  If `class` is invoked on an integer, (e.g., `8.class`) then it will return the `Fixnum` class.  The `Integer` class is a superclass of both `Fixnum` and `Bignum`.  Since this is the first Ruby diagnostic, I'll be lenient and accept both `Fixnum` and `Integer`, even if they're not properly capitalized.

lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = arr.delete_if {|item| item > 34}
```
bernardlee: A more general way of doing this is `arr.pop(2)`.

lib/diagnostic.rb
```diff
@@ -106,10 +106,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = puts "#{person[:first_name]} #{person[:age]}"
```
bernardlee: This will cause an error since `person` doesn't exist.  The answer that we're looking for is a statement that creates a hash called `person` with the requested keys and values.  `person = { first_name: 'really good name', age: 12345 }`.

lib/diagnostic.rb
```diff
@@ -106,10 +106,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = puts "My name is #{person[:first_name]}"
```
bernardlee: Assuming that `person` is defined, this will print to the console but not set it as the default return value to `person`.  It should be `person.default = "My name is #{person[:first_name]}"`.

bernardlee:
10/15

PR #155, opened 2017-02-02
responding to the diagnostic

novakw267:
I noted where I wasn't sure of things in sudo.
I referenced http://ruby-doc.org/core-2.3.0/Array.html a good amount for help

Comfort 3
clarity 2 

I might be low balling the ratings.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,23 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'print or puts'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "movie: 'star wars episode viii'"
+# I am a little confused how to present it in this form. I would normalally just say
+# movie = 'star wars episode viii'
 
 # String interpolation in Ruby
 #
@@ -29,31 +31,34 @@ Diagnostic.variable = 'your response here'
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = "{'#character'} is really a Sith Lord."
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = '%'
+#not sure on this.
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = '.div or /'
+# not sure with this
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['5 % 2', '5 / 2']
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = [==, .equals, .eql,]
+# I am unsure what you're asking for, I assume these are operators?
 
 # Flow control in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,23 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'print or puts'
```
bernardlee: We want to run a script from the terminal.  These are Ruby methods that print to the console.  `ruby lib/example.rb`.

lib/diagnostic.rb
```diff
@@ -6,21 +6,23 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "movie: 'star wars episode viii'"
```
bernardlee: This will cause an error.  We just want the name of the variable based on the given string.  Ruby's naming convention is "snake case" so this should just be `star_wars_episode_viii`.

lib/diagnostic.rb
```diff
@@ -29,31 +31,34 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = "{'#character'} is really a Sith Lord."
```
bernardlee: String interpolation in Ruby is of the form: `"#{...}"`.  This should be `"#{character} is really a Sith Lord."`.

lib/diagnostic.rb
```diff
@@ -29,31 +31,34 @@
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = '%'
```
bernardlee: Decimal numbers are represented by the `Float` class.

lib/diagnostic.rb
```diff
@@ -29,31 +31,34 @@
...
+#not sure on this.
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = '.div or /'
```
bernardlee: Integers are represented by the `Fixnum` class.

lib/diagnostic.rb
```diff
@@ -29,31 +31,34 @@
...
+# not sure with this
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['5 % 2', '5 / 2']
```
bernardlee: `5 % 2` evaluates to `1`, which is an integer.  `5 / 2` also evaluates to `1` since neither number is being converted into a floating point number and integer division is performed instead.

lib/diagnostic.rb
```diff
@@ -29,31 +31,34 @@
...
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = [==, .equals, .eql,]
```
bernardlee: `false` and `nil` are the only falsy values in Ruby.

lib/diagnostic.rb
```diff
@@ -69,18 +74,19 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'First we look to see if its batman, if not then we just
```
bernardlee: We are looking for the return value here and it's the string `The Dark Knight`.

lib/diagnostic.rb
```diff
@@ -91,13 +97,15 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = '.delete_at(2, 3)'
```
bernardlee: This method needs to be invoked on `arr` and `delete_at` only takes one argument.

lib/diagnostic.rb
```diff
@@ -91,13 +97,15 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '2, or 46'
```
bernardlee: `[12, 34, nil, nil, 99]`

lib/diagnostic.rb
```diff
@@ -106,10 +114,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = { first_name: ralph, Age: 39}'
```
bernardlee: This will cause an error since `ralph` isn't defined.

lib/diagnostic.rb
```diff
@@ -106,10 +114,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = '@person'
```
bernardlee: `person.default = "#{person[:first_name]} is my first name."`

bernardlee:
3/15

PR #156, opened 2017-02-02
diagnostic

dsweetser:
-used class notes, google, and last night's study response

Comfort: 4
Clarity: 3

did not understand one of the questions no matter how much I thought about it. Everything else was okay!

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby ./lib/example.rb
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
bernardlee: Ruby uses "snake case" for its naming convention so this should be `star_wars_episode_viii`.

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = "#{character} is really a Sith Lord."
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = "decimal numbers in ruby are called floats"
```
bernardlee: The `Float` class is specifically what we're looking for.

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```
bernardlee: Specifically, it's `Fixnum`.  `Integer` is a superclass of `Fixnum` and `Bignum`.

lib/diagnostic.rb
```diff
@@ -69,18 +69,18 @@
...
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = true
```
bernardlee: Ruby doesn't require an explicit `return`.  The most recently evaluated expression is the return value.  e.g.,

```ruby
def no_explicit_return
  1 + 1
end

p no_explicit_return # => 2
```

lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
...
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = 'I keep staring at this and not understanding what it is asking'
```
bernardlee: It's asking what `arr` looks like after the statement `arr[arr.length + 2] = 99` is executed.  `arr[arr.length + 2] = 99` sets the element at the length of the array plus two to `99`.  In this case, `arr` was originally `[12, 34, 56, 67]`.  `arr.pop(2)` changes it to `[12, 34]`.  `arr[arr.length + 2] = 99` is equivalent to `arr[4] = 99`, which sets the fifth element to `99`.  After all of this is done, `arr` is `[12, 34, nil, nil, 99]`.

lib/diagnostic.rb
```diff
@@ -106,10 +106,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = { :age => 3, :first_name => "Dan"}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person.default = "#{person[first_name]} loves alcohol, board games, and dancing."'
```
bernardlee: This will cause an error since `first_name` is not defined.

bernardlee:
11/15

PR #158, opened 2017-02-02
Diagnostic Complete.

lbreguet:
-comfort: 3
-clarity: 4
-Stackoverflow, rub-doc

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry lib/example.rb'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -6,21 +6,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry lib/example.rb'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
bernardlee: The variable should be named `star_wars_episode_viii` according to Ruby's convention of using "snake case".

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'floating object'
```
bernardlee: The `Float` class is what we're looking for.

lib/diagnostic.rb
```diff
@@ -29,31 +29,31 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```
bernardlee: The `Fixnum` class is what we're looking for, but I'll accept this since `Integer` is a superclass of `Fixnum` and `Bignum`.

lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
...
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.delete_if { |i| i > 55 }'
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '46'
```
bernardlee: `[12, 34, nil, nil, 99]`.

lib/diagnostic.rb
```diff
@@ -106,10 +106,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = [age: 20, first_name: "Louis"]'
```
bernardlee: This creates a one-element array with a hash.

lib/diagnostic.rb
```diff
@@ -106,10 +106,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person[:first_name].default = "He is pretty confused."'
```
bernardlee: This will cause an error.  Assuming that `person` is properly defined, `person[:first_name]` will be a string and `String` doesn't have a `default` method.

bernardlee:
11/15

PR #159, opened 2017-02-02
Answer questions related to Ruby

wilsonkaya:


lib/diagnostic.rb
```diff
@@ -6,21 +6,22 @@
...
+Diagnostic.script = 'ruby lib/diagnostic.rb'
+
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'irb'
```
bernardlee: We're looking for the REPL that we prefer over `irb`, which is `pry`.

lib/diagnostic.rb
```diff
@@ -6,21 +6,22 @@
...
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'best_movie = "star wars episode viii"'
```
bernardlee: We want you to name a variable after "star wars episode viii" using Ruby's naming convention (snake case).  `star_wars_episode_viii`.

lib/diagnostic.rb
```diff
@@ -29,14 +30,14 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'floating point'
```
bernardlee: The `Float` class is specifically what we're looking for.

lib/diagnostic.rb
```diff
@@ -91,13 +92,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop'
```
bernardlee: This will only remove one element from the end of `arr`.

lib/diagnostic.rb
```diff
@@ -91,13 +92,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[12, 34, 56, 99]'
```
bernardlee: `[12, 34, nil, nil, 99]`

lib/diagnostic.rb
```diff
@@ -106,10 +107,12 @@
...
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = {:age => 30, :first_name => "fatih"}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
```
bernardlee: This doesn't concatenate the value of `:first_name` in the `person` hash and a message as a default value.  `person.default = "#{person[:first_name]} is my first name."`.

bernardlee:
9/15

PR #168, opened 2017-02-02
response.

moshiko1988:
it took me longer time to understand the questions and
not get to all of them.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,20 +6,20 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -69,18 +69,18 @@
...
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = true
```
BenGitsCode: `false`

lib/diagnostic.rb
```diff
@@ -6,28 +6,28 @@
...
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
-
+movie = 'star wars episode viii'
```
BenGitsCode: just looking for snake case here `star_wars_episode_viii`

PR #169, opened 2017-02-02
ruby-diagnostic

cathyob:
answers to diagnostic

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -6,21 +6,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -29,30 +29,32 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'floats'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```
BenGitsCode: These are actually called `Fixnum` fixed numbers

lib/diagnostic.rb
```diff
@@ -29,30 +29,32 @@
...
+Diagnostic.numbers = [0.5, 7]
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
+falsy = [0, NaN, ''];
+
 Diagnostic.falsy = []
```
BenGitsCode: Make sure you put your answer in the right place. `nil, false`

PR #170, opened 2017-02-02
Answer diagnostics

AshlenPrice:
Comfort 3
Clarity 3

I used my notes and previous lessons.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -2,25 +2,25 @@ require 'ostruct'
 Diagnostic = OpenStruct.new
 
 # Running scripts from the command line
-#
+
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'Ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars = star wars episode viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -2,25 +2,25 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars = star wars episode viii'
```
BenGitsCode: That's how to assign a variable, although that wouldn't work. 

Just looking for camel case, which you had in your variable name.

lib/diagnostic.rb
```diff
@@ -29,19 +29,19 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '${charecter} is really a Sith Lord.'
```
BenGitsCode: Would be correct if not for the typo.

lib/diagnostic.rb
```diff
@@ -29,19 +29,19 @@
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = '.0001'
```
BenGitsCode: The question asked for a ruby string. `Float`

lib/diagnostic.rb
```diff
@@ -29,19 +29,19 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = '1'
```
BenGitsCode: The question asked for a ruby string. `Fixnum`

lib/diagnostic.rb
```diff
@@ -69,18 +69,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'Bruce Wayne'
```
BenGitsCode: `The Dark Knight`

lib/diagnostic.rb
```diff
@@ -69,18 +69,18 @@
...
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = true
```
BenGitsCode: False.

lib/diagnostic.rb
```diff
@@ -91,13 +91,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr = [12, 34, 56, 67].slice(-2, 2)'
```
BenGitsCode: `arr.pop(2)`  There are other ways to do this, but your solution returns those two elements, we wanted to remove them to see something like `'[12, 34, nil, nil, 99]'`

PR #178, opened 2017-04-13
Adding responses to diagnostic

jim-moody:
Clarity: 4
Comfort: 4

I would not have gotten all these without running the tests and seeing why they failed.

I don't think we learned about Fixnum in class so I had "Integer" instead of "Fixnum" but I saw the test failing so I googled it and found this resource which was helpful:
http://stackoverflow.com/questions/21372649/what-is-the-difference-between-integer-and-fixnum

I also forgot that you don't need parens when calling ruby methods so I initially had the following code for the array pop question `arr.pop(2)` , but once I saw the test failing I could see that it should be `arr.pop 2`

Everything else was pretty straightforward.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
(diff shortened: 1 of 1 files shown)


jordanallain:
15/15

💯 Way to use the tests Jim! That's what they are there for. 🔥 

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #180, opened 2017-04-13
Response

andyjzhong:
Clarity: 3
Comfort: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby ./lib/example.rb
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "movie_title = 'star wars episode viii'"
 
 # String interpolation in Ruby
 #
@@ -31,31 +31,31 @@ Diagnostic.variable = 'your response here'
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'floats'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'...?
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['3.14159', '3']
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = ['nil', 'false']
 
 # Flow control in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby ./lib/example.rb
```
jordanallain, line 11: This will work only if the `lib` directory exists in your root directory. You can leave off the `./` if you are already in the directory that `lib` is in. Also the answer was supposed to be in a string.

lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
```
jordanallain, line 18: Again we were looking for the answer to be a string but `'pry'` is the correct answer.

lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "movie_title = 'star wars episode viii'"
```
jordanallain, line 25: We wanted the variable to be `star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 34: Remember we need to use double quotes for string interpolation in Ruby.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'floats'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'...?
```
jordanallain, line 46: `Fixnum` is what we were looking for here.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['3.14159', '3']
```
jordanallain, line 51: You have your numbers in quotes so they are actually strings!

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = ['nil', 'false']
```
jordanallain, line 58: Again you can leave the quotes off here because `false` and `nil` are actual values in Ruby.

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
```
jordanallain, line 96: Remember you can also leave off the parens `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = 'arr = [12, 34, nil, nil, 99]' #This doesn't look right though. This is the result from pry but I would have expected only 4 s
```
jordanallain, line 102: Nope this is right!

lib/diagnostic.rb
```diff
@@ -108,10 +108,17 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'Andy = Person.new(':favorite_number', ':first_name')'
+
+class Person
+  def initiatialize (num, name)
+    @favorite_number = num
+    @first_name = name
+  end
+end
```
jordanallain, line 118: We were looking for the shorthand here:

`'person = {first_name: "Andy", favorite_number: 8}'`

lib/diagnostic.rb
```diff
@@ -108,10 +108,17 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'Andy[:first_name] = 'Andy''
```
jordanallain, line 124: Remember we need to use interpolation here.

jordanallain:
8/15

Good job Andy, a couple of them you were allllmost right on with but got caught up on some syntax.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #181, opened 2017-04-13
Added response to diagnostic.rb

ghost:
For ruby-diagnostic

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.' # double quotes would be needed, but that seems to mess up the rest of this doc.
```
jordanallain, line 34: I'll give you credit here because of your comment. Double quotes are necessary though, it would only mess up the document if you forgot to delete a hanging quote from somewhere else so try to debug it next time.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'Fixnum'
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['1.1', '-3']
```
jordanallain, line 51: You have these numbers as strings!

lib/diagnostic.rb
```diff
@@ -71,18 +71,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'Just your average billionaire'
```
jordanallain, line 74: Actually should be 'The Dark Knight'

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
```
jordanallain, line 96: Remember in Ruby you can also leave off the parens `arr.pop 2`.

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[12, 34, , 99]'
```
jordanallain, line 102: The array would actually have two `nil` spots where the numbers used to be.

lib/diagnostic.rb
```diff
@@ -108,10 +108,12 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = {:favorite_number => 42, :first_name => "joe"}
```
jordanallain, line 111: Remember that we can shorten these even more:

`'person = {favorite_number: 42, first_name: "joe"}`

lib/diagnostic.rb
```diff
@@ -108,10 +108,12 @@
...
+'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person.new { |hash, key| hash[key] = "#{person.first_name}" like the number #{avorite_number}"}
```
jordanallain, line 118: So this is using the new constructor, we were looking for the short hand version:

jordanallain:
11/15

Great work Joe, make sure to use the tests to your advantage, they would have nudged you in the right direction on some of the ones you missed.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #182, opened 2017-04-13
Fallon response

jonifallon:
ran out of time

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii = "something"'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 34: Careful we need to use double quotes for string interpolation in Ruby.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float (which you do by to_f)'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer (which you do by to_i)'
```
jordanallain, line 46: We were looking for `Fixnum` here.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['50.25', '3']
```
jordanallain, line 51: You wrapped these in quotes so they are actually strings here!

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
```
jordanallain, line 96: Good we can also leave off the parens in Ruby:

`arr.pop 2`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = 'arr = [12, 34, 99, 99]'
```
jordanallain, line 102: The array would actually look like: `'[12, 34, 56, 67, nil, nil, 99]'`

jordanallain:
9/15

Good job Joni, no worries about running out of time!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #183, opened 2017-04-13
Ruby Diagnostic

cpm5041:
Comfort: 4
Clarity: 3

I felt pretty good with this one. The only thing that still tripped me
up a bit was the hashes. I feel comfortable with the methods and
syntax.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby 'lib/example.rb''
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii_the_last_jedi'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby 'lib/example.rb''
```
jordanallain, line 11: You don't need the quotes around the file path.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord. <-(love this theory btw)'
```
jordanallain, line 34: Close but we need double quotes for string interpolation in Ruby.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```
jordanallain, line 46: We were looking for `Fixnum`.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['4.10', '50']
```
jordanallain, line 51: Wrapping these in quotes makes these strings not numbers.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = ['false','nil']
```
jordanallain, line 58: Again these are values in Ruby so you don't need to wrap them in quotes.

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.delete(2); arr.delete(3)'
```
jordanallain, line 96: This works only if you know what values are in the array. We can use `pop` to remove the last element from an array no matter what is there.

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person_hash = {favorite_number: 420, first_name: Sampson}'
```
jordanallain, line 111: Good but you should have wrapped `Sampson` in quotes.

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'new_hash = Hash.new { |hash, name| hash[name]= "Dr said #{name} needs a backiatomy." }'
```
jordanallain, line 117: This is using Hash.new, we wanted the shorthand syntax:

jordanallain:
9/15

Good work Chris, got snagged up on a few so check out the solution if you're still confused.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #184, opened 2017-04-13
Diagnostic completed

Ryaong792:
Fist of 5 would be a 3.  I thought i was a 4 but i got confused with the
questions and how to answer them.  But i did feel that i had a grasp
on the subject till today.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -4,25 +4,26 @@ require 'ostruct'
 Diagnostic = OpenStruct.new
 
 # Running scripts from the command line
-#
+example1 = example.new(24)
+end.
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = '#{@value}'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl =  starwars = "star wars episode vii"
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = '#{charactor} is really a Sith Lord'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


Ryaong792:
just read the answer and I feel dumb I really over thought what was wanted for the answers.

lib/diagnostic.rb
```diff
@@ -4,25 +4,26 @@
...
 
 # Running scripts from the command line
-#
+example1 = example.new(24)
+end.
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = '#{@value}'
```
jordanallain, line 12: We were looking for

`ruby lib/example.rb`

lib/diagnostic.rb
```diff
@@ -4,25 +4,26 @@
...
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl =  starwars = "star wars episode vii"
```
jordanallain, line 19: Close, should be: `star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -4,25 +4,26 @@
...
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = '#{charactor} is really a Sith Lord'
```
jordanallain, line 26: Super close here, but remember we need double quotes for string interpolation in Ruby.

lib/diagnostic.rb
```diff
@@ -38,7 +39,7 @@
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'to_i'
```
jordanallain, line 42: We just wanted the keywords that Ruby uses which are Float and Fixnum.

lib/diagnostic.rb
```diff
@@ -71,7 +72,8 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'Batman is the darknight and bruce wayne is just your average
```
jordanallain, line 75: Should be `The Dark Knight`

lib/diagnostic.rb
```diff
@@ -82,7 +84,7 @@
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = true
```
jordanallain, line 87: Actually it is `False`

jordanallain:
0/15

Hey Ryan looks like you struggled with this but I bet you just misunderstood some of the directions. Let me know if you want to go over any of these with me.

PR #185, opened 2017-04-13
Response ruby-diagnostic - Larson

rlarsonil71:
Comfort: 4
Clarity: 4

I used class notes and the various intro Ruby trainings to complete this diagnostic.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
jordanallain, line 25: Remember in Ruby we use snake_case:

`star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.slice!(arr.length-1)'
```
jordanallain, line 96: Hmm, I think this would work but why not just use `arr.pop 2`

jordanallain:
14/15

Fantastic work Ron!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

rlarsonil71:
Thanks Jordan!

PR #186, opened 2017-04-13
Ruby Diagnostic

octater:
Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 34: Careful, remember in Ruby we need to use double quotes to do string interpolation.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```
jordanallain, line 46: We were looking for `Fixnum` here.

lib/diagnostic.rb
```diff
@@ -71,18 +71,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'if since its the last expression executed ???'
```
jordanallain, line 74: It would return `'The Dark Knight'`

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person[:favorite_number] = 14'
```
jordanallain, line 111: First we need to make the Hash, it would look like this:

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = ' person = Hash.new { |hash, key| hash[key] = "#{first_name} is a super-dude." }'
```
jordanallain, line 117: This would be with the constructor but we were looking for the shorthand syntax, like this:

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
```
jordanallain, line 96: We can also leave off the parens in Ruby like so: `arr.pop 2`

jordanallain:
11/15

Good work Terry! Got snagged on a couple of them but you were on the right track.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #191, opened 2017-04-13
Diagnostic response.

danielleletarte:
Comfort: 3
Clarity: 3

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'require_relative lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'require_relative lib/example.rb'
```
jordanallain, line 11: We can actually just use `ruby lib/example.rb`

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 34: Remember we need double quotes in Ruby to do string interpolation!

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'Decimals are a flat object, which is part of the numberic class of object'
```
jordanallain, line 41: True we were looking for the keyword though in Ruby which is `Float`

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'Integers are an integer object, which is part of the numeric class of object'
```
jordanallain, line 46: Here we were looking for `Fixnum`

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['5.7', '8']
```
jordanallain, line 51: Careful you wrapped the numbers in quotes turning them into strings.

lib/diagnostic.rb
```diff
@@ -71,18 +71,18 @@
...
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = no
```
jordanallain, line 85: Heh, we wanted true or false! The answer was false so I'll give you credit for this.

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
```
jordanallain, line 96: Good we can also leave off the parens in Ruby: `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = 'Not sure what this question is asking. Without the last to elements, arr=[12,34]'
```
jordanallain, line 102: It would end up with nil in the two spots we popped and then 99 last:

`'[12, 34, 56, 67, nil, nil, 99]'`

lib/diagnostic.rb
```diff
@@ -108,7 +108,7 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = '{@favorite_number = favorite_number,  @first_name = first_name}'
```
jordanallain, line 111: Close but we don't need the @ symbol unless we are working with classes generally.

'person = {first_name: "Lauren", favorite_number: 7}'

jordanallain:
7/15

Good work Danielle. Check out the tips I gave and let me know if anything is still unclear!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #197, opened 2017-04-13
Ruby quiz

glasneed:


lib/diagnostic.rb
```diff
@@ -8,54 +8,58 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative 'lib/example.rb'
```
jordanallain, line 11: We can just use `ruby lib/example.rb`

lib/diagnostic.rb
```diff
@@ -8,54 +8,58 @@
...
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'require 'pry''
```
jordanallain, line 18: Just `pry` will get us into the Ruby REPL.

lib/diagnostic.rb
```diff
@@ -8,54 +8,58 @@
...
 
 # String interpolation in Ruby
 #
 # Use string interpolation to complete the sentence below. Replace "ANSWER" with
 # the correct code.
-
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 33: Close but remember that we need to use double quotes for string interpolation in Ruby.

lib/diagnostic.rb
```diff
@@ -8,54 +8,58 @@
...
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = '2.45'.to_f
```
jordanallain, line 40: We wanted the keyword here which is `Float`

lib/diagnostic.rb
```diff
@@ -8,54 +8,58 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = '2.45'.to_i
```
jordanallain, line 45: Here we were looking for `Fixnum`

lib/diagnostic.rb
```diff
@@ -8,54 +8,58 @@
...
+numbers[0] = an_integer.to_s
+numbers[1] = a_float.to_s
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = ['false', 'nil']
```
jordanallain, line 62: Super close but you wrapped them in quotes which makes them strings. The keywords false and nil are values in Ruby when not wrapped in quotes.

lib/diagnostic.rb
```diff
@@ -93,7 +97,11 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = '
+ets = ['cat', 'dog', 'rabbit', 'parrot']
+a.delete('parrot')
+a.delete('rabbit')
+'
```
jordanallain, line 104: This is confusing, much easier to just use `array.pop 2`

jordanallain:
5/15

Hey Conor looks like you ran out of time so check out the solution if you want and let me know if something doesn't make sense.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #198, opened 2017-04-13
Amanda's answers to Ruby study

aeabbott:
3/3- I'm feeling okay about Ruby. I still don't have the syntax down so I
have to refer a lot to my notes.

I was unclear on what the last question was actually asking for.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,6 +8,8 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
+require relative = ('lib/example.rb')
+
 Diagnostic.script = 'your response here'
 
 # Using the REPL
@@ -15,12 +17,16 @@ Diagnostic.script = 'your response here'
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
+pry
+
 Diagnostic.repl = 'your response here'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
+
+star_wars_variable = "Star wars episode viii"
 
 Diagnostic.variable = 'your response here'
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,20 +8,26 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
+require relative = ('lib/example.rb')
+
```
jordanallain, line 12: We can just use `ruby lib/example.rb`

lib/diagnostic.rb
```diff
@@ -8,20 +8,26 @@
...
 Diagnostic.repl = 'your response here'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
+star_wars_variable = "Star wars episode viii"
+
```
jordanallain, line 30: You had the right idea here so I'll give you credit but we wanted the title of the movie to be the variable: `star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -31,31 +37,32 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'fixed number'
```
jordanallain, line 53: I'll give you credit for this but technically in Ruby it is `Fixnum`

lib/diagnostic.rb
```diff
@@ -93,7 +100,7 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = arr.shift(2).shift(3)
```
jordanallain, line 103: So shift will actually remove the element from the front of the array. We want to use `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -108,10 +115,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = { first_name: 'Amanda', favorite_number: 13}
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = person[:first_name] = 'Cindy'
```
jordanallain, line 124: Super close here.

`'person.default = "#{person[:first_name]} is super cool"'`

jordanallain:
12/15

Great job Amanda, only missed a few!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #200, opened 2017-04-13
Add response

khan-ngo:


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
jordanallain, line 25: Careful here, we use snake_case in Ruby for variable names:

`star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -31,31 +31,35 @@
...
+Diagnostic.interpolation = "#{character} is really a Sith Lord."
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'Decimal(a)'
+Diagnostic.decimal = '(a).to_f'
```
jordanallain, line 42: We were just looking for `Float` here.

lib/diagnostic.rb
```diff
@@ -31,31 +31,35 @@
...
+
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = '(a).to_i'
```
jordanallain, line 48: Just wanted `Fixnum` here.

lib/diagnostic.rb
```diff
@@ -71,18 +75,19 @@
...
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
+ruby terminates upon a return
 
-Diagnostic.returns = nil
+Diagnostic.returns = ''
```
jordanallain, line 90: False was correct.

lib/diagnostic.rb
```diff
@@ -93,13 +98,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.chomp(56,67)'
```
jordanallain, line 101: So we can use `arr.pop 2` to get rid of the last two elements of an array even if we don't know how long the array is.

lib/diagnostic.rb
```diff
@@ -93,13 +98,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '101'
```
jordanallain, line 107: The new array would be `[12, 34, nil, nil, 99]`

lib/diagnostic.rb
```diff
@@ -108,7 +113,7 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'people{:favorite_number :first_name}'
```
jordanallain, line 116: This would be correct:

jordanallain:
6/15

Hey Khanh let me know if you want to go over any of these answers!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #201, opened 2017-04-13
Resonse

donpowers:


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'lib/examaple.rb'
```
jordanallain, line 11: Forgot `ruby` before the file name.

lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'irb'
```
jordanallain, line 18: We were looking for the REPL that we prefer to irb which is `pry`

lib/diagnostic.rb
```diff
@@ -31,31 +31,33 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 34: Remember in Ruby we need double quotes to do string interpolation.

lib/diagnostic.rb
```diff
@@ -31,31 +31,33 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'Float'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```
jordanallain, line 46: We call them `Fixnum` in Ruby.

lib/diagnostic.rb
```diff
@@ -31,31 +31,33 @@
...
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['10.234', '10']
```
jordanallain, line 51: When we put quotes around a number it converts it to a string so be careful there.

lib/diagnostic.rb
```diff
@@ -31,31 +31,33 @@
...
 
 # Falsiness in Ruby
 #
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
+# A non-boolean value that counts as true is called "truthy," and a non-boolean
+# value that counts as false is called "falsey."
 
-Diagnostic.falsy = []
+Diagnostic.falsy = ['string1'=='string', 1 == 2, array1[]==array2[]]
```
jordanallain, line 60: There are only two falsy values in Ruby: `[false, nil]`

lib/diagnostic.rb
```diff
@@ -93,13 +95,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop arr.pop'
```
jordanallain, line 98: Good, we can also do `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -93,13 +95,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[2, 34,99]'
```
jordanallain, line 104: There would actually be two nil values between 34 and 99.

lib/diagnostic.rb
```diff
@@ -108,7 +110,7 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person{ :favorite_number "favorite_number, :first_name "first_name"}'
```
jordanallain, line 113: Very close:

`'person = {first_name: "Don", favorite_number: 8}'`

jordanallain:
5/15

Hey Don looks like you ran out of time here. Check out the tips I left and let me know if any of them don't make any sense.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #203, opened 2017-04-13
Response

jscohen:
Clarity: 4
Comfort: 4

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -2,7 +2,7 @@
 
 require 'ostruct'
 Diagnostic = OpenStruct.new
-
+#
 # Running scripts from the command line
 #
 # In a Ruby string, write the command you would use to run a script at
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii'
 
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
```
jordanallain, line 34: Careful, we need double quotes in Ruby for string interpolation.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'Fixnum'
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['5.1', '5']
```
jordanallain, line 51: Careful with quotes, it turns your number into a string.

lib/diagnostic.rb
```diff
@@ -71,18 +71,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'The Dark Night'
```
jordanallain, line 74: Good but it should have been `Knight`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.slice!(2,3)'
```
jordanallain, line 96: We can do this even easier with `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = '{first_name: 'Jon', favorite_number: 0}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = "#{person_hash[:firstname]} + rocks!"
```
jordanallain, line 117: Super close here but when you are using string interpolation you don't need to use the `+` operator.

jordanallain:
Hey Jon looks like this PR was in really late so I marked it as late. You did well on it would have had 12 correct.

PR #206, opened 2017-04-20
Completes Ruby Diagnostic

sstone72389:
Completes required ruby diagnostic.

Comfort: 3
Clarity: 3

I used Google and my class notes.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = lib/example.rb
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = lib/example.rb
```
scottyscripts, line 12: In Ruby, we can use the `ruby` keyword to run scripts
so `ruby lib/example.rb`

lib/diagnostic.rb
```diff
@@ -108,10 +110,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = { favorite_number: 7, first_name: "spencer" }'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = Hash.new { |hash, key| hash[key] = "Key #{first_name} is now part of person hash" }
```
scottyscripts, line 119: We can set a default return value on the hash like so
`person.default = "#{person[:first_name]} is great"`

scottyscripts:
12/15 Awesome!

PR #208, opened 2017-04-20
Response

mangoceylon:
Clarity: 5
Comfort: 5

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby './lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
scottyscripts, line 26: Remember to use snake case when naming variables! 
`star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer'
```

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = arr.pop(-2)
```
scottyscripts, line 97: Close! we would pass `2` as an argument not `-2`
Pop takes a parameter in Ruby of how many elements we want to remove

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '46'
```
scottyscripts, line 103: So if we take the array from the result above we have
arr = [12,34]
arr[arr.length + 2] really is saying arr[4] = 99
So if we set arr[4] to 99 , arr[2] and arr[3] are not defined, thus becoming nil
we end up with [12, 34, nil, nil, 99] !!!

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = { favorite_number: 7, first_name: 'Jon', occupation: 'exterminator' }
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'Hello #{person[:first_name]}'
```
scottyscripts, line 117: We can set a default return value like this!
`person.default = "#{person[:first_name]} is great"`
Good job on accessing the name correctly!

scottyscripts:
10/15 GREAT!

PR #210, opened 2017-04-20
Answers to Ruby diagnostic

jbeltrami:
clarity: 4
comfort: 4

Used previous notes, google and stack overflow to provide my answers.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,21 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'read_in_english = "star wars episode viii"'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -114,4 +114,4 @@
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person = Hash.new { |hash, person| hash[person] = "Sorry, #{person[:first_name]} does not exists"}'
```
scottyscripts, line 117: you could set the default return value by `person.default = "#{person[:first_name]} is great"`

lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'read_in_english = "star wars episode viii"'
```
scottyscripts, line 26: `star_wars_episode_viii`
I think you misread the question but get the idea of it

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'integers'
```
scottyscripts, line 42: Float

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'instance object'
```

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
 
-Diagnostic.numbers = ['decimal', 'integer']
+Diagnostic.numbers = ['120', '120']
```
scottyscripts, line 52: These are actually technically both strings `'120'` instead of `120`
an example of a decimal would be `1.02` or something like that

scottyscripts: -.5 no decimal

lib/diagnostic.rb
```diff
@@ -108,7 +108,7 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = Hash.new { | hash, key | hash[key] = {favorite_number[7], first_name["joao"]} }'
```
scottyscripts, line 112: `person = {first_name: "Lauren", favorite_number: 7}`

scottyscripts:
10.5/15 Awesome!

PR #211, opened 2017-04-20
Add responses to Ruby Diagnostic

nriser:


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative './lib/example.rb'
```
scottyscripts, line 12: We can run scripts in Ruby with the `ruby` keyword
so `ruby 'lib/example.rb'`

lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii'
```
scottyscripts, line 26: Remember to use snake case when naming variables!
star_wars_episode_viii

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'array.pop.pop'
```
scottyscripts: Close, but this isn't doing what it looks like it's doing.
When we call pop on array, the value returned is the last element (a FixNum)
when we try to call the second pop, it's not being called on the array, but rather what was last returned
this will throw an error because Fixnum has no pop method!
You could do 
`array.pop
array.pop
`
or more idiomatic, `array.pop 2`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[12, 34]'
```
scottyscripts, line 103: `[12, 34, nil, nil, 99]` is the answer here!

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = '{ favorite_number: 4, first_name: 'Nikki'}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = "#{favorite_number} is the favorite number of #{first_name}"
```
scottyscripts, line 117: You could set the default return value by `person.default = "#{person[:first_name]} is great"`

scottyscripts:
10/15 Great job!

PR #214, opened 2017-04-20
answers questions

Anddrew914:


lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby example.rb
```
scottyscripts, line 12: Specify `lib/example.rb` unless you are in the lib Directory

lib/diagnostic.rb
```diff
@@ -8,21 +8,21 @@
...
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = i_hope_this_one_is_good = 'star wars episode viii'
```
scottyscripts, line 26: `star_wars_episode_viii` would read like "star wars episode viii" in English
You did demonstrate that you knew how to instantiate a variable though! 
-.5

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'integer' ? I'm not really sure of the question here
```
scottyscripts, line 47: in Ruby, every object has a type! For integer numbers, their object type is `Fixnum`
https://en.wikibooks.org/wiki/Ruby_Programming/Data_types

lib/diagnostic.rb
```diff
@@ -71,18 +71,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'Just your average billionaire'
```
scottyscripts, line 75: Here it would actually be "The Dark Knight" 
The if statement checks if `batman` is true. Since we defined it above it is true and that block will execute

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
...
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.pop(2)'
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '[12, 34, nil, 99]'
```
scottyscripts, line 103: So Close! It would be `[12, 34, nil, nil, 99]`
Remember that arr[0] is the first element in the array!

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person { favorite_number: 1, first_name: Bob,}'
```
scottyscripts, line 111: Don't forget = sign after `person`
Good!

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person.default = "#{person[:'first_name'] makes great burgers}"
```
scottyscripts, line 117: Nice!

scottyscripts:
11/15 Great job!!

PR #215, opened 2017-04-20
diagnostic

eliottenos:
Comfort: 3
Clarity: 3

My brain is still turning on --___--

stackoverflow, ruby docs, zetcode

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,14 +8,14 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require 'lib/example.rb'
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = const repl = require('repl')
 
 # Naming conventions in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,14 +8,14 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require 'lib/example.rb'
```
scottyscripts, line 12: We can execute scripts with the `ruby` keyword
so `ruby 'lib/example.rb'`

lib/diagnostic.rb
```diff
@@ -8,14 +8,14 @@
...
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = const repl = require('repl')
```
scottyscripts, line 19: You would just enter `pry` to enter our favorite Ruby REPL!
A REPL is an acronym for Read–Eval–Print Loop or fancy words for a thing where we can interact with the language and code

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = character + 'is really a Sith Lord.'
```
scottyscripts, line 35: Nice! This is definitely a valid way of interpolating strings in Ruby!
A more common way would be `"#{character} is really a Sith Lord."`

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = float
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = instance
```
scottyscripts, line 47: `Fixnum`

lib/diagnostic.rb
```diff
@@ -93,7 +93,7 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = arr.delete(2, 3)
```
scottyscripts, line 97: You could totally use the delete method, but not like this. 
A more common way would be `arr.pop 2`
Try getting it to work with the delete method also if you want!
https://ruby-doc.org/core-2.4.1/Array.html#method-i-delete_if

scottyscripts:
7/15 These are tough to do in the morning (or any time of day when you're learning so many new concepts!) Take a look at my feedback and feel free to reach out to one of us if you have any questions

PR #216, opened 2017-04-20
Pull request for ruby-diagnostic

noobiwankenoobi:
Confident in most of the answers
Except the last two because I ran out of time

Comfort: 4
Clarity: 4

Some of them I knew from the lessons and studies, and everything else I found using Google (the documentation)

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,23 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = ruby lib/example.rb
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = pry
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = star_wars_episode_vii
+
+# convention is to use underscores and lowercase
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -31,31 +33,32 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{:character} is really a Sith Lord.'
```
scottyscripts, line 37: The only thing here is we don't need the colon.
This would actually print "character is really a Sith Lord."

lib/diagnostic.rb
```diff
@@ -71,18 +74,20 @@
...
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'The Dark Knight'
 
 # In a Ruby string, write what keyword you use for "else if" clauses in Ruby.
 
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = elseif
+
```
scottyscripts, line 82: `elsif` with no `e`
Don't confuse with JavaScript's `else if`
-.5

scottyscripts: Ahhh just saw your comment. Yeah in Ruby there is no `e` and it's one word

lib/diagnostic.rb
```diff
@@ -93,13 +98,15 @@
...
+Diagnostic.remove_from_array = arr.pop(2)
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = '12, 34, nil, nil, 99'
+
+# is this just arr[4] = 99 ?
```
scottyscripts, line 109: That's exactly what it is ^^ 
Correct! but you forgot the brackets around your array

lib/diagnostic.rb
```diff
@@ -108,10 +115,14 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = '{person: [favorite_number: 42, first_name: ross]}'
+
+# I don't think this is right, needed more time
```
scottyscripts, line 121: Close! Hint: you don't need the array in there

lib/diagnostic.rb
```diff
@@ -108,10 +115,14 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default =
+
+# more time
```
scottyscripts, line 128: `person.default = "#{person[:first_name]} is awesome"`

scottyscripts:
11.5/15 Great! Try out those last two, I think you can get them!

PR #217, opened 2017-04-20
answers to diagnostic

joebelmonte:


joebelmonte:
Minor note, but this question:

_Does ruby require an explicit return from methods? Answer true or false._

Isn't really worded as a t/f.  I'd suggest either changing it to "True or False: Ruby requires an explicit return from methods." Or changing the response options to yes or no.

lib/diagnostic.rb
```diff
@@ -31,31 +31,31 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = "#{character} is really a Sith Lord."
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'BigDecimal'
```
scottyscripts, line 42: `Float`

lib/diagnostic.rb
```diff
@@ -93,13 +93,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.delete_if {|num| num > 50 }'
```
scottyscripts, line 97: Nice!! This is one way of doing it! 
You could also do `arr.pop 2` (which you will probably see more often in the wild)

lib/diagnostic.rb
```diff
@@ -108,10 +108,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = {favorite_number: 33, first_name: 'John'}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = Hash.new { |hash, key| hash[key] = "We have added #{key} to the list." }
```
scottyscripts, line 117: You can set the default return value like
`person.default = "#{person[:first_name]} is awesome"`

scottyscripts:
13/15 Awesome!
Thanks for the note!

PR #218, opened 2017-04-20
Diagnostic.rb response

Pouk91:
Comfort: 5
Clarity: 3

Use resources provided in diagnostics

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,21 +8,22 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = 'ruby example.rb' #Assuming user is in lib folder
 
 # Using the REPL
 #
 # In a Ruby string, write the command you would use to enter the REPL we
 # prefer over `irb`
 
-Diagnostic.repl = 'your response here'
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "star wars episode viii"
+           puts variable
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,21 +8,22 @@
...
+Diagnostic.repl = 'pry'
 
 # Naming conventions in Ruby
 #
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = "star wars episode viii"
+           puts variable
```
scottyscripts, line 27: We can define that variable by using snake case `star_wars_episode_viii` and it would be read as "star wars episode viii"

lib/diagnostic.rb
```diff
@@ -31,19 +32,19 @@
...
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = '#{character} is really a Sith Lord.'
 
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = ''
```
scottyscripts, line 43: `Float`

lib/diagnostic.rb
```diff
@@ -31,19 +32,19 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = ''
```
scottyscripts, line 48: `Fixnum`

lib/diagnostic.rb
```diff
@@ -71,18 +72,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'Just your average billionaire'
```
scottyscripts, line 76: it would actually be `'The Dark Knight'`
we are checking if batman is true, and it is since we defined it above!

lib/diagnostic.rb
```diff
@@ -71,18 +72,18 @@
...
-Diagnostic.else_if = 'your response here'
+Diagnostic.else_if = 'elsif'
 
 # Returns in Ruby
 #
 # Does ruby require an explicit return from methods? Answer true or false, and
 # replace nil.
 
-Diagnostic.returns = nil
+Diagnostic.returns = 'Yes'
```
scottyscripts, line 87: `false`

lib/diagnostic.rb
```diff
@@ -93,13 +94,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'remove_from_array.pop(2)'
```
scottyscripts, line 98: it would be `arr.pop(2)`

lib/diagnostic.rb
```diff
@@ -108,10 +109,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = { :favorite_number => "", :first_name => ""}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person[:first_name] = "Hi my name is Pinaro"'
```
scottyscripts, line 118: You are accessing the value correctly, but how would you concatenate it?

scottyscripts:
6/15

PR #219, opened 2017-04-20
Responses

JLisaW:
I'm not sure I interpreted the questions correctly.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,7 +8,7 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative 'lib/example.rb'
 
 # Using the REPL
 #
@@ -22,7 +22,7 @@ Diagnostic.repl = 'your response here'
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii.'
 
 # String interpolation in Ruby
 #
@@ -31,8 +31,9 @@ Diagnostic.variable = 'your response here'
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = Hash.new { |hash, key| hash[key] = '#{key} is really a Sith Lord.'
 
+Diagnostic.interpolation[:character]
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
@@ -43,7 +44,7 @@ Diagnostic.decimal = 'your response here'
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = object.to_i
 
 # Write an example of a decimal and an integer in Ruby. Replace the Ruby strings
 # in the following array with your examples.
@@ -55,7 +56,7 @@ Diagnostic.numbers = ['decimal', 'integer']
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = [nil]
 
 # Flow control in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,7 +8,7 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = require_relative 'lib/example.rb'
```
scottyscripts, line 12: we can execute scripts using the `ruby` keyword
so `ruby 'lib/example.rb'`

lib/diagnostic.rb
```diff
@@ -22,7 +22,7 @@
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star wars episode viii.'
```
scottyscripts, line 26: remember we use snake case when naming a variable in Ruby.
we could define it as `star_wars_episode_viii`

lib/diagnostic.rb
```diff
@@ -31,8 +31,9 @@
 
 character = 'Jar Jar Binks'
 
-Diagnostic.interpolation = 'ANSWER is really a Sith Lord.'
+Diagnostic.interpolation = Hash.new { |hash, key| hash[key] = '#{key} is really a Sith Lord.'
 
+Diagnostic.interpolation[:character]
```
scottyscripts, line 36: We can interpolate variables into strings using `#{}` with Ruby
So to interpolate our variable `character` we could do 
`"#{character} is really a Sith Lord."`

lib/diagnostic.rb
```diff
@@ -43,7 +44,7 @@
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = object.to_i
```
scottyscripts, line 48: In Ruby, the Object type of integers is `Fixnum`

lib/diagnostic.rb
```diff
@@ -55,7 +56,7 @@
 # Write all the values that evaluate to "falsy" in Ruby. Write these values in
 # a Ruby array.
 
-Diagnostic.falsy = []
+Diagnostic.falsy = [nil]
```
scottyscripts, line 60: Correct! nil will evaluate to falsy.
`false` will also evaluate so

lib/diagnostic.rb
```diff
@@ -93,13 +94,15 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array =
+arr.delete[2]
+arr.delete[3]
```
scottyscripts, line 100: Here you would use the `pop` method
`arr.pop 2`

lib/diagnostic.rb
```diff
@@ -93,13 +94,15 @@
...
 
 # Taking the result from the question above, what would `arr` now equal:
 arr[arr.length + 2] = 99
 # Write your response as a ruby string
 
-Diagnostic.array_add_to = 'your answer here'
+Diagnostic.array_add_to = [12, 34, 56, 67, nil, 99]
```
scottyscripts, line 106: So close! Remember that arrays start at index 0 (`arr[0]` is the first element)
So it would be `[12, 34, 56, 67, nil, nil, 99]`

lib/diagnostic.rb
```diff
@@ -108,10 +111,11 @@
 # Use shorthand to make this one line long.
 # Write your response as a Ruby string.
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = {:favorite_number, :first_name}
```
scottyscripts, line 115: Close! You just forgot to instantiate the hash with the values
`person = {first_name: "Lauren", favorite_number: 7}`

lib/diagnostic.rb
```diff
@@ -108,10 +111,11 @@
...
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = Hash.new('person')
+person[:first_name] +
```
scottyscripts, line 121: `person.default = "#{person[:first_name]} is awesome!"`

scottyscripts:
3/15 . One of the trickiest part of development is translating english to code and vice versa! Take a look at the feedback I gave, and reach out to myself or another consultant if you need more help with this!

PR #221, opened 2017-04-20
Time up

d00medman:
Had trouble with default value, repl question and unsure of what falsy means.

### DIFF
--- a/lib/diagnostic.rb
+++ b/lib/diagnostic.rb
@@ -8,7 +8,7 @@ Diagnostic = OpenStruct.new
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = pry lib/example.rb
 
 # Using the REPL
 #
@@ -22,7 +22,7 @@ Diagnostic.repl = 'your response here'
 # In a Ruby string, write what you would write to properly name a variable in
 # Ruby that read in English as "star wars episode viii".
 
-Diagnostic.variable = 'your response here'
+Diagnostic.variable = 'star_wars_episode_viii'
 
 # String interpolation in Ruby
 #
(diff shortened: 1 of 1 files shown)


lib/diagnostic.rb
```diff
@@ -8,7 +8,7 @@
 # In a Ruby string, write the command you would use to run a script at
 # `lib/example.rb`
 
-Diagnostic.script = 'your response here'
+Diagnostic.script = pry lib/example.rb
```
scottyscripts, line 12: here you would use 'ruby lib/example.rb'

lib/diagnostic.rb
```diff
@@ -33,22 +33,24 @@
...
 
+"#{character} is really a Sith lord"
+
 # Numbers in Ruby
 #
 # What is the type of object in Ruby for decimal numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.decimal = 'your response here'
+Diagnostic.decimal = 'float'
```
scottyscripts, line 44: Correct! you will usually see object types in ruby capitalized, so 'Float'

lib/diagnostic.rb
```diff
@@ -33,22 +33,24 @@
...
 
 # What is the type of object in Ruby for integer numbers? Write your answer in
 # a Ruby string.
 
-Diagnostic.integer = 'your response here'
+Diagnostic.integer = 'int'
```
scottyscripts, line 49: 'Fixnum' is the correct answer

lib/diagnostic.rb
```diff
@@ -71,18 +73,18 @@
 
 # What value will be returned. Write your response as a Ruby string.
 
-Diagnostic.flow_control = 'your response here'
+Diagnostic.flow_control = 'Just your average billionaire'
```
scottyscripts, line 77: the answer is 'The Dark Knight' because 'batman' is true since we defined it above

lib/diagnostic.rb
```diff
@@ -93,13 +95,13 @@
 # In a Ruby string, write how you would remove the last two elements from this
 # array
 
-Diagnostic.remove_from_array = 'your answer here'
+Diagnostic.remove_from_array = 'arr.slice(-2, arr.length)'
```
scottyscripts, line 99: This will return the last 2 elements in the array, but it would not remove them
You could remove them by saying `arr.pop 2`

lib/diagnostic.rb
```diff
@@ -108,10 +110,10 @@
...
 
-Diagnostic.person_hash = 'your answer here'
+Diagnostic.person_hash = 'person = {first_name: "Alex", favorite_number: 7}'
 
 # Set a default return value to your `person` hash that concatenates the value
 # of person[:first_name] and a message about them.
 # Write your response as a Ruby string.
 
-Diagnostic.hash_default = 'your answer here'
+Diagnostic.hash_default = 'person = Hash.new { |hash, key| hash[key] = "Go Fish: #{person[:first_name]}" }''
```
scottyscripts, line 119: Here it's
`person.default = "#{person[:first_name]} is awesome"` (of course you could use any sting at the end

scottyscripts:
7/15 Take a look at the feedback and let myself or another consultant know if you're still having trouble.
Falsy is just another term for false, with 'truthy' used for true!

